PR #19, opened 2015-12-11
Add skill experience related operators to #calc.

jp8900308:
E - current skill exp
N - exp to next level
Z - last skill exp event amount
Q - general exp to level conversion, opposite of L

### DIFF
--- a/calc.c
+++ b/calc.c
@@ -8,6 +8,7 @@
 #include "calc.h"
 #include "hud.h"
 #include "platform.h"
+#include "stats.h"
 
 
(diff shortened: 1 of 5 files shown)


pjbroad:
Hi.  Thanks for the patch, it merges clean and looks fine to me.  One minor point is that I get compiler warnings from the strlen(.....shortname) operations.  This just need some casting but I thought you might prefer to clean up the patch before I commit it.

pjbroad:
Thanks for updating, and thanks for your contribution!

PR #21, opened 2016-01-04
Map Editor function: Fine Camera Rotation

feeltheburn:
Holding Shift button while rotating camera allows fine adjusting of camera angle.


feeltheburn:
Could probably be cleaned up, but it works.

A lot of detailed work in map editor requires very slight camera movement for a very specific angle, and the normal camera rotation is incapable of that. 

With this, holding down Shift will allow very specific camera placement.

PR #39, opened 2017-02-08
Alternative dialog systems in game launcher

gvissers:
After reinstalling eternal lands on a new (linux) computer, I found that the game launcher on linux uses zenity for its dialogs. It is a nice system, and the dialogues created are pretty, but I don't have it installed on this box and I don't really want to either. I can imagine more people have been bitten by this. So while it's certainly possible to work around (either install zenity, or start the game directly), let's try to be a bit more user friendly and support more dialog systems.

This patch series introduces support for (terminal) dialog and kdialog in the game launcher. The script uses the first dialog command it can find, checking in order zenity, kdialog, and dialog(*),

Support is not perfect: while both systems (**) somewhat support aligning multiple columns in the menu items of the server selection screen, I can find no way to add column headers as is done in the zenity launcher. Also, the default size of the server selection screen in kdialog is too small, with no apparent way to change it from the command line. Still it is useful to me at least, and it seems to work.

(*) straight console could also be done, in fact all that's needed is to hack up a text-based server selection screen.
(**) kdialog seems to be heavily inspired by dialog, terminology and capabilities are very similar as far as I can see.

### DIFF
--- a/3d_objects.c
+++ b/3d_objects.c
@@ -108,7 +108,7 @@ void draw_3d_object_detail(object3d * object_id, Uint32 material_index, Uint32 u
 
 	//debug
 
-	if (object_id->self_lit && (!is_day || dungeon) && use_lightning) 
+	if (object_id->self_lit && (!is_day || dungeon) && use_lightning)
 	{
 		glColor3fv(object_id->color);
 	}
@@ -222,7 +222,7 @@ void draw_3d_object_detail(object3d * object_id, Uint32 material_index, Uint32 u
 	}
 	else
 	{
-		glDisable(GL_TEXTURE_2D);		
+		glDisable(GL_TEXTURE_2D);
 	}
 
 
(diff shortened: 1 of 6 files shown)


pjbroad:
IIRC (I wrote this stuff a long while ago), the zenity dialogue is only used by the Ubuntu / Debian packages and the zenity package is a dependency in the control file; so it should have installed by default.  So it will work unless you override the dependencies, no one should get bitten.

That said, the patch is probably a good thing to do and the dependency the packages have could be changed to allow for the options.

gvissers:
Arch uses it as well (patching paths to executables and data files), with an /optional/ dependence on zenity. Which is somewhat reasonable, since you can start the game without the launcher. Personally, I don't think it's an invasive patch (though for obvious reasons it's untested with zenity :)), but feel free to ignore it if you don't think it's worth the trouble.

pjbroad:
Oh.  Didn't know about the Arch usage.  If you have not tested with zenity then I'll do that before merging.  It may be a week or so before I have time though.

PR #62, opened 2019-04-02
Update interface.c

erdies:
This is to enable 2k monitors for fullscreen. Custom resolution are rejected if they are not in the list.


pjbroad:
This will be a useful change, thanks for doing it.  However, you have changed the order of the display modes which unfortunately, means that anyone with that mode enabled in their el.ini file will get the wrong mode when starting the client.  Could you change the pull request to add the new mode at the end of the list.  Also, perhaps you could add the 4k mode too 3840 x 2160.

erdies:
Thanks for pointing to this, will do it!

erdies:
Changes are done. I added the 4k mode, too.

pjbroad:
Thanks.  All merged.

PR #76, opened 2020-05-11
Various macOS fixes

bendoughty:
A few macOS-specific fixes and an updated Xcode project file. Project file commit looks messy due to the change in format over the years.


pjbroad:
Thanks.

PR #92, opened 2020-12-07
Fix cal_render_bones that is broken when DEBUG is enabled

brunoramoslu:
Fix cal_render_bones that is broken when DEBUG is enabled (to test make sure do disable AVP/animation program).


gvissers:
Merged, thanks for catching this.

PR #101, opened 2021-01-02
Add tooltips in map editor toolbox buttons

brunoramoslu:
I was wondering if we could add tooltips to the map editor buttons, something like this.

I could try to rework the code a bit to prevent duplicating the "positions" of of the buttons if needed.

### DIFF
--- a/elconfig.c
+++ b/elconfig.c
@@ -2992,6 +2992,7 @@ void init_vars(void)
 	add_var(OPT_BOOL,"show_position_on_minimap","spos",&show_position_on_minimap, change_var, 0,"Show Pos","Show position on the minimap",HUD);
 	add_var(OPT_SPECINT,"auto_save","asv",&auto_save_time, set_auto_save_interval, 0,"Auto Save","Auto Save",HUD,0,INT_MAX);
 	add_var(OPT_BOOL,"show_grid","sgrid",&view_grid, change_var, 0, "Show Grid", "Show grid",HUD);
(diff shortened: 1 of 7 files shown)


feeltheburn:
Tooltips are quite annoying once you know the tools, and there aren't many to learn here.

If added, there needs to be an option to not show them. (Simplest method being a variable set in the mapedit.ini file.)

gvissers:
As Burn seems to be the only one actually using the map editor, I'd take his comments into account :smile: 

Also, since you're cleaning up code, you might want to address a pet peeve of mine: you could take the opportunity to remove the hard-coded pixel boundaries and replace them by dynamic values in an array (which needs to be initialized at startup). This would help in adding UI scaling to the map editor (the buttons are really small on my system). Either that, or at least replace them by appropriately named #define constants. Then use the same values in the click handler.

brunoramoslu:
I will clean up the code a bit and add an option to enable/disable tooltips.

I'm also thinking about addind the 't' key to control tooltip being shown or not.

I think setting the tool tip to activate by default should be ok, if you can later disable it.

feeltheburn:
That sounds fine, as long as there's a way to deactivate, by variable or by click. ;)

brunoramoslu:
I have done a couple of changes to improve the pull request:

- CTRL+t to enable/disable the tooltips
- Option in mapedit.ini to disable tooltips
- Some refactoring to improve the code (also in preparation for further changed in the toolbar)

Any feedback would be welcome :slightly_smiling_face:

gvissers:
I think this works well enough, so I have merged it. One suggestion for further a possible improvement: try adding a semi-transparent  background behind the tooltips, so that they're easier to read.

I'll take a stab at scaling the buttons.

PR #105, opened 2021-01-04
Fix DROP_ITEM message_length in quickbar

brunoramoslu:
Fix DROP_ITEM message_length in quickbar

When doing a ctlr + left click to drop all from the quick bar the the **DROP_ITEM** message length was set to 4 causing item quantity to misbehave.
The fix sets the **DROP_ITEM** message length to 6 like all the other DROP_ITEM messages in the code.

State before drop all from quick bar (ctrl + left click)
inventory-quickbar-before-drop-all-in-quickbar
State after drop all from quick bar (ctrl + left click)
inventory-quickbar-bag-after-drop-all-in-quickbar


gvissers:
It's a bloody miracle the server accepted the short count. Good find! Merged, and thanks.

PR #111, opened 2021-01-09
Fix background environment sounds

gamil-zirak:
Update map id's in mapsfx xml files so that the correct map's environment sounds are loaded.

http://www.eternal-lands.com/forum/index.php?/topic/61493-i-found-a-servere-client-sound-bug/


pjbroad:
Thanks for the fix.

PR #113, opened 2021-01-11
fix drawing boundaries for debug map sound

gamil-zirak:
When turning on -DDEBUG_MAP_SOUND, the area boundaries defined in mapsfx files are not being drawn correctly in the tab map screen.  This change scales the boundaries drawn on the map so that they match the size of the map and draws the boundaries in the correct locations.


pjbroad:
Thanks for the fix.

PR #116, opened 2021-01-19
fixed transposed x,y coordinates for irsis south beach background sound

gamil-zirak:
Original file had the x and y values switched for one of the points in the Irsis south beach sound boundaries. This change fixes that and the sound now plays as expected.


pjbroad:
Thanks for the fix.

PR #120, opened 2021-03-10
Various macOS fixes (preliminary p9 edition)

bendoughty:
A few macOS-specific fixes and an updated Xcode project file.


pjbroad:
Thanks for the pull request.  Is there a way to define the the path to nlohmann_json during the build process rather than in the code source file?

bendoughty:
I should have known my original method wasn't necessarily the best way to go about doing that. I wasn't sure how to set the path correctly in Xcode and had concerns about the portability of the project since we've had some interest from others looking to compile their own client.

I have done some research on how to set the paths correctly within the project file, and the issue is now fixed (along with some other non-portable paths which were still lingering from the old days).

Thank you for the prompt, it has proven to be a good learning opportunity. :-)

pjbroad:
Thanks again for the contribution and for the update.

PR #124, opened 2021-04-08
Map sounds

gamil-zirak:
Update map sound config files to use the map file name rather than the map array index number.
Update the client code to read map sound config files that use either the map file name or the map array index number.  Keeping the functionality to read map sound config files that use map array index number is for backwards compatbility with older map sound config files.


pjbroad:
Thanks for the patch.

pjbroad:
Thanks for the patch,

PR #125, opened 2021-04-13
Filter the just exp message in console

gamil-zirak:
When mixing in schools, hide the just exp message in console when the mixed item filter is enabled.


pjbroad:
Thanks for the feature.

PR #141, opened 2021-07-19
SSL support for android

gvissers:
This is a merge of gvissers/ssl into the android branch, together with a small fix and an update to Android.mk to compile with USE_SSL.


pjbroad:
Thanks for getting this working.  To keep align with the main branch, I merged the SSL code from their and just too the Android.mk updates from this pull request.

PR #145, opened 2021-08-17
ranging exp/arrow calculation

gamil-zirak:
Fix ranging exp/arrow calculation when ranging exp and total shots are out of sync.


gvissers:
Merged, thanks!

PR #149, opened 2021-08-27
Mouse position and markings in the map window

gamil-zirak:
Improve the accuracy of the map markers in the map window and map click to walk.


gvissers:
Merged, thanks!

PR #152, opened 2021-09-09
Xcode project update

bendoughty:
Here's an updated Xcode project file which enables the new SSL features for macOS users. I also took the opportunity to do a bit of housekeeping and make the project easier to navigate, since there has been some interest lately.

I anticipate another small change will be required prior to the next release (if only to bump the version number), but for now this seems to be building and running great on both newer (ARM64) and older (x86_64) Macs.

### DIFF
--- a/macosx/Eternal Lands.xcodeproj/project.pbxproj
+++ b/macosx/Eternal Lands.xcodeproj/project.pbxproj
@@ -1715,18 +1715,13 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_CXX_LANGUAGE_STANDARD = "c++14";
-				CLANG_ENABLE_OBJC_WEAK = YES;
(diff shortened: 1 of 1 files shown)


gvissers:
Merged, thanks.

Quick question though: I see you have C++ standard set to C++17. So far, we have made a deliberate effort to stick to C++11, is there any reason  the XCode project sets C++17 (and if so, should we fix it? :smile: )

bendoughty:
No, no reason (certainly not an intentional one on my part, at least). It looks like that happened this morning after I updated to the latest Xcode beta (to test macOS Monterey support) and found it spitting out various warnings about new things Apple doesn't like this year. Would you prefer for this to be reverted, or does it not matter so much?

gvissers:
Nah, it's no big deal, I was just wondering if it was a conscious decision, and whether we should act upon it.  Just leave it as is, if that is what works best for you.

PR #156, opened 2021-09-21
Android

gvissers:
An issue was raised in the forums (http://www.eternal-lands.com/forum/index.php?/topic/61508-new-development-version-of-the-android-client/&do=findComment&comment=597504) about the client not running on old versions of Android. The problem is that the client has the OpenSSL libraries libcrypto.so and libssl.so dynamically linked in, and when the client starts the system version of these libraries are already loaded. These system libs are too old, and miss functions used by the client. Fix this by linking the OpenSSL libraries statically, forcing the client to use the bundled libs.

Apologies for the messy merge, really only the client Makefile has changed.


pjbroad:
Thanks for the update.

pjbroad:
Its time to delete your feature branch and start a new one when things get that out of date. :)

gvissers:
> Its time to delete your feature branch and start a new one when things get that out of date. :)

In retrospect, that was a better idea. And perhaps things would've gone better if I had rebased instead of trying to merge. Anyhow, thanks for merging, and glad to see in the forums that the issue is indeed solved.

PR #167, opened 2022-01-03
Fix water reflection on Windows/Virtualbox/Intel graphics

gvissers:
Fix a number of issues with the water reflection running the client on a Windows client in Virtualbox:
* The framebuffer texture is interpreted upside down by the VirtualBox driver. This is an issue that has been
encountered in other programs as well, notably on Intel integrated graphics chipsets. Fix this by detecting
if this is the case by generating a known texture and checking its contents, and flipping the scene vertically
if it occurs.
* The OpenGL implementation seems to have trouble with the water shader using a vectored multiply-add-assign
operation. Explicitly constructing a new vec2 from separate components seems to solve the issue somehow. Though I
am no expert in GLSL, the existing code certainly looks valid, so this seems to point to another driver bug.

Posting this as a pull request, so that people may check if it does not have any unwanted side effects. As far as I can tell, there should be none: the first fix should not affect properly behaving systems, and the second just rewrites the same assignment in separate components. I have tested this also in Linux, which was working correctly before, and have observed no change in behaviour. Nevertheless, better get some other testing in before we break an existing setup.

### DIFF
--- a/init.c
+++ b/init.c
@@ -1117,6 +1117,8 @@ void init_stuff(void)
 	skybox_init_gl();
 	popup_init();
 
+	check_flip_fbo_texture();
+
 	DO_CHECK_GL_ERRORS();
 	LOG_DEBUG("Init done!");
 }
--- a/reflection.c
+++ b/reflection.c
@@ -52,6 +52,68 @@ int water_buffer_usage = 0;
 int water_buffer_reflectiv_index = 0;
 int water_shader_quality = 0;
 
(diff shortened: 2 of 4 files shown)


pjbroad:
It does not appear break the nice water ripples and reflections for my main machine Linux + nvidia; which is good.  I'll check on a few other machines when I have time, including and ancient AMD laptop running Windows which does work on currently.

gvissers:
Great, thanks for testing!

PR #173, opened 2022-01-15
Try to scale and center objects inside the 3d object browser of the map editor

brunoramoslu:
There's something that has been bothering me for awhile now regarding the map editor 3d object browser.
The objects are not always visible, or not centered, in the 3d object browser.
From what I could understand from the code, the behavior depends on the size of the object.

I'm trying to play around with a scale factor to place the object inside the "canvas" always centered and scaled to fit the 3d browser window.

This is not finished yet, but I would like some comments if you have any :slightly_smiling_face: 
Also if you can test and confirm you see no issue it would be great.

Once we agree what would be the best approach to improve this I can rework the pull request.

### DIFF
--- a/map_editor/3d_objects.c
+++ b/map_editor/3d_objects.c
@@ -142,6 +142,124 @@ void draw_3d_object(object3d * object_id)
 	CHECK_GL_ERRORS();
 }
 
+#define SCALE_FACTOR 4
+void draw_3d_object_in_browser(object3d * object_id)
+{
+    int i;
+    float s_plane[4], t_plane[4];
+    float x_pos,y_pos,z_pos;
+    float x_rot,y_rot,z_rot;
+    float scale;
+
+    //fprintf(stderr, "Max Size: %f\n", object_id->e3d_data->max_size);
+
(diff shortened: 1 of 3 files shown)


gvissers:
I agree, it is an annoying issue.

And I'm sorry, I haven't looked at this in much detail yet, but the first question that came to my mind was: what does `draw_3d_object_in_browser()` do different from `draw_3d_object()`, and can the issue not be solved more easily by appropriate glScale and/or glTranslate around the calls to `draw_3d_object()` in the browser window?

Also, my feeling is that it is more of a translation issue than a scaling issue. The objects are drawn in the upper half of the fields, leaving the lower half nearly empty aside from the object name. My gut feeling is that by lowering them a bit, most of them can be made to fit easily. But as I said, I haven't looked at it to closely.

brunoramoslu:
Thank you for your comments, really helpful.

The code change was really an ugly rough approach to show that items can be aligned using this approach.

I'll refactor the code an bit and try to clean it up.
It should be better now, but it's still not ready.

PR #180, opened 2022-03-12
Fix for GL-related macOS compilation issues

bendoughty:
Recent work on the water shader caused a few GL-related compilation issues for Mac users. It was discovered that these issues stemmed primarily from the use of an old cobbled-together elglext.h file, rather than the standard glext.h shared by the other platforms. This PR fixes this issue.


pjbroad:
Thanks for the fix.

PR #186, opened 2022-03-26
Android client branch merge into main

pjbroad:
Merge of the Android branch.
* All Android only changes #def using ANDROID.
* Additional Android clean-ups tagged with ANDROID_TODO
* Changes not using #def are meant to be only intentional fixes that were initially done on the android client; mostly scaling.

I will do a squash merge later in case anyone wants to review and comment.


gvissers (approved):
Some nitpickery added as comments, by I don't think there's anything that will break the game. I will admit I haven't done the Android build itself yet, from looking at the patches I think there might be some compiler warnings on usused functions. But as said, havent tried.

pjbroad:
Wow that is a lot of nit picking. I'll be sure to pay the same attention to your next work.

pjbroad:
End of the start for Android.

PR #211, opened 2024-05-28
Add `-ld_classic` linker flag so build succeeds on Xcode 15

### LINKED ISSUE #210: Macos build action fails with macos-latest (now 14)
Hi @bendoughty and @lukehorvat.  I've just pushed some some minor changes and the Macos build action is broken.  I suspect that this is because the action uses macos-latest and that has changes from 12 to 14 since the last build.  Any idea's how to fix it? Other than pinning it to 12 ;)

lukehorvat:
Fixes #210.

The latest macos image uses Xcode 15, which introduces a new linker: https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Linking

I added the `-ld_classic` linker flag recommended there to revert back to the old linker and it seems to fix things. We'll need to drop this flag at some point though, since it sounds like it will be removed in a future Xcode release.

@pjbroad @bendoughty

### DIFF
--- a/macosx/Eternal Lands.xcodeproj/project.pbxproj
+++ b/macosx/Eternal Lands.xcodeproj/project.pbxproj
@@ -1828,6 +1828,7 @@
 					"-lc",
 					"-lstdc++",
 					"-liconv",
+					"-ld_classic",
 				);
 				SDKROOT = macosx;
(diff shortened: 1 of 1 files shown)


pjbroad:
Thanks for the fix @lukehorvat .

PR #215, opened 2025-02-26
Slider widget for options window

bendoughty:
Here’s a PR for a new “slider” widget type I have been working on. It is intended for use in the options window as the first part of an effort to make that area of the interface less cluttered and more intuitive for the user.

As is typically the case with UI sliders, the absolute value of the option is now no longer directly shown to the user within the interface. To make this a little more palatable to those who may miss it, I have added an extra line to the right-click context menu which displays the current value of the option. This change applies to slider options only.

I have updated some candidate options to use the new option type, these are very conservative changes based on Human Interface Guidelines from Gnome and Apple (thanks for the guidance, @pjbroad!). If the new widget type is approved I am happy to make adjustments to these selections as discussion progresses 😊

It would be useful if android folks could have a play around with the new widget, as I unfortunately do not possess a device capable of running the android client.

Screenshot 2025-02-26 at 08 19 07

https://github.com/user-attachments/assets/2cb3f1f1-c227-4b63-aca9-9ec29bdd0e4e

### DIFF
--- a/elconfig.c
+++ b/elconfig.c
@@ -2134,6 +2134,7 @@ static int context_option_handler(window_info *win, int widget_id, int mx, int m
 	{
 		case OPT_INT:
 		case OPT_INT_F:
+        case OPT_INT_S:
 			set_var_OPT_INT(option->name, (int)new_value);
 			break;
 		case OPT_MULTI:
@@ -2144,6 +2145,7 @@ static int context_option_handler(window_info *win, int widget_id, int mx, int m
 			set_var_OPT_BOOL(option->name, (int)new_value);
 			break;
 		case OPT_FLOAT:
+        case OPT_FLOAT_S:
 			set_var_OPT_FLOAT(option->name, new_value);
 			break;
 		default:
@@ -2160,6 +2162,7 @@ static int add_cm_option_line(const char *prefix, var_struct *option, float valu
 	{
 		case OPT_INT:
 		case OPT_INT_F:
+        case OPT_INT_S:
 		case OPT_MULTI:
 		case OPT_MULTI_H:
 			safe_snprintf(menu_text, sizeof(menu_text), "\n%s: %d\n", prefix, (int)value);
@@ -2168,6 +2171,7 @@ static int add_cm_option_line(const char *prefix, var_struct *option, float valu
 			safe_snprintf(menu_text, sizeof(menu_text), "\n%s: %s\n", prefix, ((int)value) ?"true": "false");
 			break;
 		case OPT_FLOAT:
+        case OPT_FLOAT_S:
 			safe_snprintf(menu_text, sizeof(menu_text), "\n%s: %g\n", prefix, value);
 			break;
 		default:
--- a/elconfig.c
+++ b/elconfig.c
@@ -2117,13 +2117,28 @@ static int context_option_handler(window_info *win, int widget_id, int mx, int m
 {
 	float new_value = 0;
 	var_struct *option = (var_struct *)cm_get_data(cm_id);
-	if (menu_option == 1)
-		new_value = option->default_val;
-	else if (menu_option == 2)
-		new_value = option->config_file_val;
-	else
-		return 1;
-
+    
+    // Bump the menu_option index for slider option types as their context menu has an additional line
+    if(option->type == OPT_INT_S || option->type == OPT_FLOAT_S)
+    {
+        if (menu_option == 2)
+            new_value = option->default_val;
+        else if (menu_option == 3)
+            new_value = option->config_file_val;
+        else
+            return 1;
+    }
+    // All other types
+    else
+    {
+        if (menu_option == 1)
+                new_value = option->default_val;
+            else if (menu_option == 2)
+                new_value = option->config_file_val;
+            else
+                return 1;
+    }
+    
 	switch (option->type)
 	{
 		case OPT_INT:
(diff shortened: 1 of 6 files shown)


pjbroad:
Nice work @bendoughty, I really like it.  A few initial comments:

1. There is an issue if you have "User Interface Scaling Factor" anything other than 1.0.  The scaling of the widget in the window is fine, but the slider does not move to the correct position when clicked, and lags when sliding.  For > 1.0 value you can get to the most right side.
2. The code works on Android (apart form the above issue), but:
 -  the keyboard is opened when you click.  You should just be able to remove the ANDROID #defs that call SDL_StartTextInput() in the el_config.c handler functions.
 - the scaling is too wide, overlapping the option text.  I tested removing the 2 x scaling just for the width and start position, keeping the 2 x height.  That looks fine to me.
3. I see you have use space characters rather than TABs for indent throughout the patch.  The convention for EL is TABS.  Though not 100% follow, it is particularly an issue in code that is does use TABS as the space using code shows non-aligned.  PITA I know, but hopefully your text editor supports fixing this easily.

bendoughty:
Thank you for taking the time to test out the widget, I really appreciate the constructive feedback 😊

1. The integer truncation bug causing the slider positioning issue should now be fixed. I have tested scaling up, scaling down, and restarting the client at various scaling levels.

2. I am glad the widget is (mostly) working as expected for Android, but clearly I need to take a look at picking up a cheap device I can use to test things like this in future. The keyboard issue you noted has been resolved, however I think it might be best for me to tap you in (if that’s okay) re: the scaling issue as I am a little dubious about making changes blindly.

3. Sorry about the formatting problem, I have a new machine and it would appear I was blissfully unaware that Xcode now defaults to indenting with spaces(?!) - this should now be resolved. 

4. I have fixed the compiler warning, thank you for pointing it out. 

5. The “current value” and other context menu option lines for slider floats have now been limited to 2 significant figures. 

6. I have added a fix which I believe *should* resolve the context menu issue for Android - let me know.

pjbroad:
Thanks for making the changes, looks good.

The following patch fixes the Android with scaling.   It also replaces the cm_destroy() call with the poorly named cm_post_show_check() that close the context menu window rather then destroy it.  We could remove the #def around it too. [android-patch.txt](https://github.com/user-attachments/files/19143256/android-patch.txt)

One other thing I realised while testing, other widgets use sound for actions.  What do you think about adding sounds for the slider?  At least for clicks, but what about dragging.  For the Gnome desktop, you get sounds for clicks, and a sound when you stop dragging.   Both could potentially be based on mouse up events but there's nothing the the window/widget framework to provide such events; add that would be a fair sized change.

If you apply the Android patch, and assuming we pick up the sound thing later, are you ready for this to be merged?

elconfig.c
```diff
@@ -4161,6 +4325,35 @@ static void elconfig_populate_tabs(void)
 				widget_set_OnKey(window_id, widget_id, (int (*)())spinbutton_onkey_handler);
 				widget_set_OnClick(window_id, widget_id, spinbutton_onclick_handler);
 			break;
+			case OPT_INT_S:
+				label_id = label_add_extended(window_id, elconfig_free_widget_id++, NULL,
+					current_x, current_y, 0, elconf_scale, (char*)var->display.str);
+				widget_width = slider_width;
+				widget_id = slider_add_extended(window_id, elconfig_free_widget_id++, NULL,
+#ifdef ANDROID
+					window_width - right_margin - 2.0f * widget_width, current_y, 2.0f * widget_width, 2.0f * line_height,
```
pjbroad, line 4334: As discussed, the with and x position scaling should be removed to avoid making the widget too wide.

elconfig.c
```diff
@@ -4161,6 +4325,35 @@ static void elconfig_populate_tabs(void)
...
+				widget_set_OnClick(window_id, widget_id, slider_onclick_handler);
+				widget_set_OnDrag(window_id, widget_id, slider_ondrag_handler);
+			break;
+			case OPT_FLOAT_S:
+				label_id = label_add_extended(window_id, elconfig_free_widget_id++, NULL,
+					current_x, current_y, 0, elconf_scale, (char*)var->display.str);
+				widget_width = slider_width;
+				widget_id = slider_add_extended(window_id, elconfig_free_widget_id++, NULL,
+#ifdef ANDROID
+					window_width - right_margin - 2.0f * widget_width, current_y, 2.0f * widget_width, 2.0f * line_height,
```
pjbroad, line 4349: As discussed, the with and x position scaling should be removed to avoid making the widget too wide.

elconfig.c
```diff
@@ -3735,6 +3827,68 @@ static int spinbutton_onclick_handler(widget_list *widget, int mx, int my, Uint3
...
+
+static int slider_ondrag_handler(widget_list *widget, int mx, int my, Uint32 flags)
+{
+	if(widget != NULL) {
+		int i;
+		slider *button;
+		
+#ifdef ANDROID
+		if (cm_valid(CM_INIT_VALUE))
+			cm_destroy(CM_INIT_VALUE);
```
pjbroad, line 3869: As discussed, suggest replacing if () cm_destroy() with cm_post_show_check(1)

elconfig.c
```diff
@@ -3735,6 +3827,68 @@ static int spinbutton_onclick_handler(widget_list *widget, int mx, int my, Uint3
...
 
+static int slider_onclick_handler(widget_list *widget, int mx, int my, Uint32 flags)
+{
+	if(widget != NULL) {
+		int i;
+		slider *button;
+		
+#ifdef ANDROID
+		if (cm_valid(CM_INIT_VALUE))
+			cm_destroy(CM_INIT_VALUE);
```
pjbroad, line 3838: As discussed, suggest replacing if () cm_destroy() with cm_post_show_check(1)

elconfig.c
```diff
@@ -2151,14 +2220,23 @@ static void call_option_menu(var_struct *option)
 	cm_set_colour(cm_id, CM_GREY, 201.0/256.0, 254.0/256.0, 203.0/256.0);
 	cm_set(cm_id, option->name, context_option_handler);
 	cm_grey_line(cm_id, 0, 1);
+	add_cm_current_value_line(cm_options_current_str, option, *(float*)option->var);
+	cm_grey_line(cm_id, 1, 1);
```
pjbroad, line 2224: The cm_grey_line(cm_id, 1, 1) is harmless at this stage but could be moved to the add_cm_current_value_line() function or make conditional on the return value.

bendoughty:
Thank you very much for the patch, I have applied it and also made the change in call_option_menu that you recommended 😊

Re: an action sound for the slider, this is actually something I have also been wondering about. Although a perfect implementation is not currently possible, I agree that it is better to at least have something in place for clicks. I have added a confirmationary do_click_sound to the onclick handler. This results in a "click!" when the slider is clicked (and, weirdly, at the beginning of a drag).

If you are happy with this then I am happy for the merge to go ahead!

pjbroad:
All merged.  Thanks very much for adding this function.

PR #219, opened 2025-07-30
Rename scale to anim_scale

lukehorvat:
Another small rename I hope you'll consider.

From what I can ascertain, actors may have several different scales in their def files: `actor_scale`, `mesh_scale`, `bone_scale`, and `scale`.  My PR concerns this last one, `scale`. The current naming is a bit ambiguous and the purpose of it is not clear. At first I thought it might be associated with `actor->scale`, but no, that's dynamic scaling that is applied via `ADD_NEW_ENHANCED_ACTOR`.

After digging through the code a bit I can see this `scale` is only used in 3 places: for scaling the Cal3D animation keyframe translations. So it's an _animation_ scale. For this reason, I'd like to suggest renaming it to `anim_scale`. It'd be nice to rename the `scale` elements in the xml files too.

**Fun fact:**
For animations it's typically best to use the skeleton scale (`bone_scale`) or you risk the mesh becoming deformed when animating, so I questioned why EL has this separate animation scale in the first place. Looking at the actor def files, I found that big spiders are the _only_ actors with an animation scale that is different from the skeleton scale. Their skeleton is much smaller than the mesh, which results in the mesh becoming a bit deformed when animating. Here's their walking animation slowed down:

https://github.com/user-attachments/assets/baccfc4b-7b7a-46ae-bc73-adc3f51b4a82

Increasing the skeleton size to match the mesh results in a much nicer animation and mesh: 

https://github.com/user-attachments/assets/b00abce8-5a2c-4432-8405-956a7ecfb976

I was going to make a PR removing the redundant animation scale entirely in favor of skeleton scale, but then when I set the animation speed back to normal I saw why the deviation was needed for big spiders – the walking animation simply doesn't look good when the skeleton is properly sized:

https://github.com/user-attachments/assets/9357a627-9efe-4133-a88f-acbaef2c5336

So unfortunately because of spiders the separate animation scale is still needed.

### DIFF
--- a/actors.c
+++ b/actors.c
@@ -489,7 +489,7 @@ static void draw_actor_banner(actor *actor_id, const actor *me, float offset_z)
 	glGetDoublev(GL_PROJECTION_MATRIX, proj);
 	glGetIntegerv(GL_VIEWPORT, view);
 	// Input adjusted healthbar_y value to scale hy according to actor scale
-	gluProject(healthbar_x, healthbar_y, healthbar_z * actor_id->scale * actors_defs[actor_id->actor_type].actor_scale + 0.02, model, proj, view, &hx, &hy, &hz);
+	gluProject(healthbar_x, healthbar_y, healthbar_z * get_actor_scale(actor_id) + 0.02, model, proj, view, &hx, &hy, &hz);
 
 	//Save World-view and Projection matrices to allow precise raster placement of quads
 	glPushMatrix();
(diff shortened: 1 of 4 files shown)


actors.c
```diff
@@ -489,7 +489,7 @@ static void draw_actor_banner(actor *actor_id, const actor *me, float offset_z)
 	glGetDoublev(GL_PROJECTION_MATRIX, proj);
 	glGetIntegerv(GL_VIEWPORT, view);
 	// Input adjusted healthbar_y value to scale hy according to actor scale
-	gluProject(healthbar_x, healthbar_y, healthbar_z * actor_id->scale * actors_defs[actor_id->actor_type].actor_scale + 0.02, model, proj, view, &hx, &hy, &hz);
+	gluProject(healthbar_x, healthbar_y, healthbar_z * get_actor_scale(actor_id) + 0.02, model, proj, view, &hx, &hy, &hz);
```
lukehorvat, line 492: I also found a couple of places where `get_actor_scale` can be called. Included that in the PR too.

